PR #4, opened 2023-07-23
Add Scrollama; prep doc layout for Scrollama

jimjam-slam:
- Adds scripts for Scrollama.js and the polyfill
- Injects Scrollama, polyfill and our scroller init script during the Lua filter
- Makes a stub scroller init script


sidebar-cr.qmd
```diff
@@ -39,9 +48,13 @@ Diam maecenas ultricies mi eget mauris pharetra et ultrices. Ullamcorper morbi t
 
 This is a scatterplot
 
+:::{}
 Second paragraph.
 
-Third paragraph
+These two paragraphs I want to keep together. But it looks like the nested div is breaking the lua script...
```
andrewpbray, line 54: Breaking how?

jimjam-slam: Whoops I my first swing at paragraph grouping, I had some invalid markdown and the layout broke (not the Lua script 😅🥲). And then I fixed the problem but forgot to change the text in the sample doc 🤣 This is all good now I believe!

Issue #13, opened 2023-08-07
transitions from revealjs


andrewpbray:
There seem to be two very useful transitions implemented in revealjs that could be useful in the cr format: auto-animate and code highlights.

1. **Auto Animate**: [Docs](https://revealjs.com/auto-animate/) / [Source](https://github.com/hakimel/reveal.js/blob/master/js/controllers/autoanimate.js)
2. **Step-by-step code highlights**: [Docs](https://revealjs.com/code/#step-by-step-highlights) / [Source](https://github.com/hakimel/reveal.js/blob/92ee97fbfe74239e8bc196b0954bd19a23410fd4/plugin/highlight/plugin.js)

In an ideal world, we could add a revealjs dependency, then in our source / filter add whatever classes are needed to "fool" the revealjs code into thinking that scrolling an object into the viewer is analogous to making a given slide (.section) visible. But that seems like it'd be too good to be true? Alternatively, we could harvest the revealjs source for parts, modify them to our needs, then add it to our existing scrollama dependency. To add to the mix, code highlighting (sans transitions) is 

As we wade into this thicket of the revealjs library and the implementation in Quarto, it might be useful to loop in a dev like Christophe to get his advice.

Thoughts?

jimjam-slam:
Yeah, I think it makes sense to reach out! My implementation of the sticky stack is already based on `r-stack`, so I see a lot of the parallels there. I'm starting to think (just outlined in https://github.com/quarto-lab/close-read/pull/11#issuecomment-1668704230) that our API might need an adjustment or two anyway, so I think it makes sense to look at our scope broadly and evaluate right now anyway 😄

andrewpbray:
Hmmm... continuing the line of thinking around the parallels with revealjs...

### Similarities between cr and reveal

`r-stack` is used when to position .fragments on top of one another. The docs [describe fragments](https://revealjs.com/fragments/)...

> The default fragment style is to start out invisible and fade in. This style can be changed by appending a different class to the fragment.

So the `.fragment` class is very similar to our `.cr-sticky` class that's inside a `r-stack` layout. With our current transition setup, our "fragments", too, start out invisible and fade in. Animation between fragments is organized by building an index of fragments, as we do with cr-sticky objects (this is used, too, for code-highlighting). That's all to say, I wonder how much we'd buy by just appending the fragment class inside our filter.

### Differences between cr and reveal

#### Trigger- or object- based transitions

[...]

PR #15, opened 2023-08-15
Cr sidebar transition and focus

jimjam-slam:
Just setting up the PR for this forthcoming refactor!


jimjam-slam:
In [76d12b5](https://github.com/quarto-lab/close-read/pull/15/commits/76d12b58528e380a10c1dfb497b835e724d179ad)..[2acec20](https://github.com/quarto-lab/close-read/pull/15/commits/2acec2054d33c0416c9423dcc029e7c2faf0e70e) I've switched over to using an attribute instead of a class to drive interactions, and I've also injected an OJS variable for tracking the step.

In [a0b9012](https://github.com/quarto-lab/close-read/pull/15/commits/a0b901202bfab8d87d08ee479db039ce21497cc6) I've add turned on progress updates and added that to the OJS too (so you can get a percentage of the block that has crossed).

I've also started experimenting with moving from pushing the narrative pars apart with margin to using padding. The reason for this is to fight the problems with fast scrolling: because the pars are such small targets, it's very easy after a quick scroll to land somewhere outside an edge intersection and not get any update. That fixes itself as soon as you continue to scroll and hit one, but I'd prefer to get a quicker update, and I think the solution is to make the "content" of the `.cr-crossfade` to be larger (using padding).

The downside of this will be when we come 'round to mobile layout and we want the narrative blocks to have some background colour to be readable against the sticky content—you don't want that extra padding to be shaded. We may need to:

1. enforce a rule that only top-level blocks can be `.cr-crossfade`, not spans or nested blocks; and
2. similar to the sticky content, have the Lua script wrap the content in a parent block (if it isn't already) and move `.cr-crossfade` to that top level. That way, the wrapper, which triggers state updates, has all the padding, and the user's inner content is what actually gets styled.

andrewpbray:
Sorry for going dark there - first week of classes! I'm catching up on #13 and this PR right now, trying to get back into the headspace...

andrewpbray:
This looks fantastic!

Switching from margin to padding makes a lot of sense to give a bigger target to the viewer. That's a good point about the implications for the overlay layout and your solution - bumping the `.cr-crossfade` class to a parent div - makes sense to me.

**Setting scrollama options via metadata**
I love the debug overlay that's activated on this branch. I can imagine it being very helpful to an author that's fine tuning a doc. I'm thinking through what it'd take to make this setable via document metadata. I'm imagining something like this:

```yaml
---
format:
  close-read-html:
    debug: true
---
```

It seems like it'd be straightforward to add the progress bar box to the doc via the lua filter - just tacking that `.counter` div that you have onto the end of a doc when `debug: true`. Getting the horizontal lines seems more challenging as it seems like that's called within the setup of the scroller.

```js
  scroller
    .setup({
      step: ".cr-crossfade",
      offset: 0.5,
      progress: true,
      debug: true
    })
```

I imagine we might also want to allow the author to set, say, `offset: 0.4` in document metadata, so the general question is: how do we set scrollama options via metadata?

I'm puzzled as to how we'd do this. My first cut is that the lua filter would need to take the metadata and modify the js script itself but that sounds messy. I don't really know how the evaluation of the js is working in the doc, but could we modify the script to use something like `offset: my_offset` and `debug: my_debug`, where `my_offset` and `my_debug` are variables defined in the lua filter and made available to the scrollama js script via an ojs block? 

As an aside, I just wandered across this: <https://github.com/yihui/rolldown>. Neat to see he tackled the same challenge.

jimjam-slam:
The Lua script has a Quarto function down the bottom to inject an HTML dependency — I wonder if there's something similar to inject some metadata into the `<head>`. Or maybe e we just use the filter to write some extra HTML into the body. I'm pretty sure the filter has access to the frontmatter, and once the config is in the HTML we can pull it out in the JS 😊

andrewpbray:
Do you think `quarto.doc.include_text(in-header, text)` would do the trick? ([docs](https://quarto.org/docs/extensions/lua-api.html#includes))

And what format do should `text` take here? That is, what sort of markup does it need so that the JS can pick out those variables?

jimjam-slam:
Yep, I run that's exactly what we want. Technically the `text` could be any valid HTML tag (we could inject frontmatter as attributes, or we could have child elements with attributes, etc).

My instinct would be to use [`<meta>`](https://developer.mozilla.org/en-US/docs/Learn/HTML/Introduction_to_HTML/The_head_metadata_in_HTML) tags in the `<head>` with `name` and ` content` attributes for the key-value pair, since that's kinda what they're designed for—we just need to put some sort of name space on the name attributes like `cr:` to avoid collisions with things like OpenGraph.

I'd be interested to see what over Quarto extension developers do for this!

jimjam-slam:
A quick survey via [Awesome Quarto]():

* [reveal-auto-agenda](https://github.com/andrie/reveal-auto-agenda/tree/main/_extensions/reveal-auto-agenda) doesn't inject any HTML, but its Lua script _does_ read custom frontmatter using a top-level `auto-agenda` key
* [quarto-animate](https://github.com/mcanouil/quarto-animate) reads custom frontmatter and uses `quarto.doc.include_text` with the `"in-header"` position argument to pass those options in as CSS variables. It constructs the CSS (essentially CSS variables scoped to the document root) as a string, wraps it in `<style>` HTML tags, and passes the whole string into the document header

andrewpbray:
Interesting. Ok, I'll give it a go this morning, trying to at least inject custom frontmatter options into the html file if not into the js. Can the js read arbitrary html in as variables? If so, attributes within the `<meta>` tag seem like the most straightforward way to store them to me.

jimjam-slam:
Thanks!  Yeah, as long as the meta elements have unique names that we can search for, I should be able to pick them up in the JS! I'm imagining we'll need one meta element for each option we want to pass.

We could also pass them as attributes in a single div in the body or something, but then we need to make sure the Div isn't displayed.

andrewpbray:
Great! So, for example, if the yaml read:
```yaml
---
mykey1: true
mykey2: 17
---
```
Would we write to html
```html
<meta mykey1="true">
<meta mykey2="17">
```
or
```html
<meta name="mykey1" content="true">
<meta name="mykey2" content="17">
```

jimjam-slam:
I believe it would be the latter! Here’s the MdN reference for it:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta

andrewpbray:
Alright, I think i've got it working. You can check it out in #16 . Note it doesn't really do anything 😬 , it just creates `<meta cr-debug-mode="true">` in the header of the html so it can be pulled into `scroller-init.js`.

jimjam-slam:
Nice! I'll try to hook up the JS end tonight.

Re. the exact HTML shape to use for passing, there's some nice discussion
here: https://stackoverflow.com/q/10533764/3246758

*James Goldie*
Data journalist + researcher
Email: ***@***.***
Phone: +61 3 9903 2102 <61399032102>, +61 421 747 208 <+61421747208>
Web: jamesgoldie.dev
Twitter: @jimjam_slam <http://twitter.com/jimjam_slam>

PR #16, opened 2023-08-29
Add debug mode via custom yaml

andrewpbray:
The goal of this PR is to add the scrollama debug mode as a toggleable feature via document yaml. So far this branch:

1. moves yaml frontmatter from the example qmd into `_extension.yml` so that we can use the more compact `format: closeread-html`.
2. changes the name of the format in several places from `close-read` to `closeread`. It looks like the naming convention of custom formats is to have no spaces or hyphens.
3. adds a function to the lua filter that looks for a new `debug-mode: <true or false>` yaml option in the document and injects it into the header as `<meta cr-debug-mode="<true or false>">.


jimjam-slam:
This is working now! I'll merge it in and continue work on the [cr-sidebar-transitionandfocus](https://github.com/quarto-lab/close-read/tree/cr-sidebar-transitionandfocus) branch.

PR #17, opened 2023-09-08
Wrap steps + mobile overlay view

jimjam-slam:
This PR adds two things:

## 1. Wrap steps

Narrative blocks that are also steps/triggers (currently defined by `.crossfade`) are wrapped in divs, and any CR-related classes or attributes get moved to the parent div.

This solves a few problems related to sizing and fast scrolls: previously, steps that only took up one line of text weren't triggering properly at narrower screen widths (even if they were completely visible), and when a fast scroll ended inside a sidebar section but not with any particular step intersecting the offset line, the sticky wouldn't update. Now that steps take up more room on either side, they're a lot more resilient to fast scrolls and small step content (although not entirely resilient to the former: a quick scroll that ends in an interlude may still fail to trigger an update below).

There's one big user change with this: it doesn't work with spans. All steps have to be blocks. We could probably write extra logic into the Lua to handle spans, but it would still be assuming that we're only talking about spans that take up their entire parent `<p>` (ie. that a Quarto user is just writing a par wrapped in a span). Would love to hear your thoughts too, @andrewpbray, but my feeling is that it's just easier to tell the user to use divs and forget the span syntax altogether for spans.

The other downside to this is that blocks trigger a bit early now (see point 2 below to visualise!), as we have padding both above and below them. There're a few ways around that:

1. Adjust the offset to taste (say, make it trigger at 25% from the top rather than at 50%).
2. Offset in scrollama terms refer to the fraction of the screen height at which an element intersection triggers an update. But separately, we also track 'progress' (fraction of the element's height that has passed the offset). We could build progress in too, so that typical updates only happen when, say, 30% of the triggering element has progressed past the offset).

## 2. Debugging styles

I've also attached a `.cr-debug` class to our sidebars when Scrollama's debug mode is on, and added some styles for our steps so that it is easier to see the intersections happening!

### DIFF
--- a/_extensions/closeread/closeread.lua
+++ b/_extensions/closeread/closeread.lua
@@ -36,8 +36,15 @@ function make_sidebar_layout(div)
     narrative_blocks = div.content:walk {
       traverse = 'topdown',
       Block = function(block)
+        -- return only the non-sticky blocks...
         if not is_sticky(block) then
-          return block
+          -- but also wrap the ones that are steps in a div
+          if block.classes ~= nil and
+            block.classes:includes("cr-crossfade") then
+            return wrap_step(block)
+          else
+            return block
+          end
         else
           return {}
         end
@@ -81,6 +88,33 @@ function shift_class_to_block(block)
   end
   
   return block
+end
+
+-- wrap_step: wraps blocks with a .cr-crossfade (or potentially another 'step'
+-- class in future) in a div that allows us to have steps visually 
+function wrap_step(block)
+  
+  -- first extract the cr-* attributes
+  local attributesToMove = {}
+  for k, v in pairs(block.attributes) do
+    if k:find("^data-cr-") or k:find("^cr-") then
+      if block.type == "Span" then
+        quarto.log.output("Close Read warning: do not use Spans as steps!")
+      end
+      table.insert(attributesToMove, {k, v})
+      block.attributes[k] = nil
+    end
+  end
+
+  -- now do .cr-* classes: add to parent block class list and remove from
+  -- current block list
+  local classesToMove = block.classes:filter(
+    function(c) return c:find("^cr-") ~= nil end)
+  block.classes = block.classes:filter(
+      function(c) return c:find("^cr-") == nil end)
+  
+  -- finally construct a pandoc.div with the new details and content to return
+  return pandoc.Div(block, pandoc.Attr("", classesToMove, attributesToMove))
 end
 
 
--- a/_extensions/closeread/custom.scss
+++ b/_extensions/closeread/custom.scss
@@ -62,8 +62,39 @@
   }
 }
 
+/* mobile (overlay) view */
+
+@include media-breakpoint-down(lg) { 
+  .columns.cr-sidebar {
+    flex-direction: column-reverse;
+
+    .column.sidebar_col {
+      width: 100% !important;
+      padding-inline: 7.5%;
+      align-self: center;
+      z-index: 1;
+
+
+      /* wip - this is supposed to target our narrative blocks to make them readable over the sticky content, but since non-step narrative blocks (ie. bare content) also gets spaced, this breaks the spacing. we may need to wrap other blocks too, not just the steps. */
+      & > *:not(section, .cr-crossfade),
+      section > *:not(.cr-crossfade),
+      .cr-crossfade > * {
+        padding: 10px;
+        background-color: #fafafabb;
+        max-width: 90%;
+        margin-inline: auto;
+      }
+    }
+
+    .column.sticky_col {
+      width: 100% !important;
+      align-self: center;
+    }
+  }
+ }
+
 // debug
 .cr-debug .cr-crossfade {
-  background-color: yellow;
+  background-color: #ffff0099;
   border: 1px solid orange;
 }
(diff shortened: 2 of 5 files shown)


jimjam-slam:
In [f84e2f9](https://github.com/quarto-lab/close-read/pull/17/commits/f84e2f98d5555462e512c4fe18259cd720273a48) I've also had a first go at a mobile "overlay" view. It's currently set to kick in below [Bootstrap's `lg` breakpoint](https://getbootstrap.com/docs/5.0/layout/breakpoints).

I've made some default aesthetic decisions that I assume we'd want to open up to the user. I also think that the CSS is becoming complex enough that it might be difficult for others to customise unless we feed YAML options through. I think ideally users would be able to target classes with their own CSS.

You'll also see that the spacing is a little broken at the moment in the mobile view—namely, the non-step narrative blocks lose their vertical padding. I added a function to the Lua script to wrap the steps in divs so that we could put the vertical spacing on the wrappers and just cover the actual step content itself in a background colour. In hindsight, that wrapping function should probably be applied to the non-step narrative blocks too!

Anyway, see what you think!

andrewpbray:
### Wrap Steps

This seems like a fix to the fast-scroll issue! Tried as I might, I wasn't able to break it :)

A few things I'm wondering about:

*Relationship between lua functions*
Our two lua functions, `shift_class_to_block` (which should be renamed `shift_crid_to_block` I guess) and `wrap_steps`, are pretty similar. The first looks to see if there's a `cr-id` attribute on an inline element of a block and moves it to be an attribute of a div wrapping that block. The second looks for cr classes and attributes on a block and moves it to be a class of a div wrapping that block.

Could we use the approach of the first to allow, in this class, step classes to be put on inline elements and we'd still be pulling them out to the outer block? At that point, could we just combine the two functions into one, and then provide a second argument to specify what you're searching for? Probably not necessary, but might help keep track of things.

*Padding and offset*
I think adjusting the offset to taste would work just fine. Feels like that's a number that you'd probably want to fiddle with anyhow to get it just right.

One other idea: would switching from `padding` to `padding-bottom` help?

*Figure captions*
As I was fiddling around with the example qmd, it struck me to add some captions to the cr-id elements (see latest commits). Interestingly, it worked a-ok for the code-generated figure, but led to errors on both the inline and block images. Not sure what's up there but I'll take a look next block of time I have, probably Monday!

andrewpbray:
Ok, I think I figured out the deal with captions: one of them isn't being detected as sticky.

This one is detected as sticky just fine:

```markdown
:::{cr-id="pink"}
![This is a caption](pink-desktop.png)
:::
```
This one, however, is not detected as sticky:
```markdown
![This is a caption](blue-desktop.png){cr-id="blue"}
```
The first form comes in as a Div block with `cr-id` on the attributes list and the content as a Figure block (it looks like it's the Figure node type that has the `caption` key). That is: Div* (block) > Figure (block) > Plain (block) > Image (inline), where * is where the `cr-id` is.

The second form comes in as a Figure block with nothing on the attributes list. Instead, it's stored at Figure (block) > Plain (block) > Image* (inline). If you remove the caption, it interestingly parses as Para (block) > Image* (inline). So to accommodate the caption, it wraps it in a Figure block, which buries the `cr-id` one level too deep to be detected by `is_sticky`.

`is_sticky` is currently written to check whether a block is `cr-id` or one of it's Inline elements is `cr-id`. Pandoc pulls apart all nested blocks and runs them as separate blocks (so there's replications of the leafs of the tree). We have it set to `topdown` so it starts with the longest branches first, and is set to stop processing child blocks if it finds a `cr-id`.

The easy fix is to just disallow the one Inline `cr-id` syntax that we had supported: `{cr-id=""}`. Feels like a bit of a bummer since that's one of the most common things that people would append a `cr-id` to. At the same time... I have a hard time right now thinking of a comprehensive solution that doesn't have the feeling of something that'd open us to way more problems.

The solution that sounds best to me right now... I can just extend `is_sticky` allow a third case: if the block is Figure > Plain > Image, it will look for `cr-id` on the image. This was the exception to the cr-id-on-block-elements rule that we intiially carved out. This just ensures that that exception will persist if a caption is added (and the whole bit gets tossed into a Figure block).

What do you think?

jimjam-slam:
Ahh man, that's a tricky one. I really prefer the idea of having one syntax in our docs if another isn't consistently supported, but I see what you mean—tacking a caption on an image _is_ going to be one of the most common use cases.

Do you happen to know if a code chunk that outputs an image (or even a few images) and has a caption gets stickied properly? I suppose we haven't really tested those kind of computational outputs with captions much yet!

jimjam-slam:
I think my feeling is that if `is_sticky` covers images with the inline syntax, it should probably also cover figures—that's better than having one covered and not the other. I think I'd be willing to accept the steps being stricter (no inlines at all) than the stickies as a small inconsistency!

jimjam-slam:
Makes sense! I ought to be able to get the merge done today. Are you happy for me to merge it without a check from you if you don't have time for one?

I hope positconf goes well—say hi to Carlos for me! I won't be able to make the sessions live because of the time difference, but I'm keen to catch up on the recordings!

andrewpbray (approved):
I removed the inline image caption (it prevented that demo from rendering) and I renamed it to something a bit more descriptive.

Looks good to me!

PR #19, opened 2023-10-24
Revise examples

andrewpbray:


andrewpbray:
I figured a fine way to test where we're at in this extension is to try to port the "base quarto" version of the [gg close read](https://github.com/andrewpbray/gg-close-read/blob/master/gg-close-read.qmd) document. This was largely an exercise in deleting, which was great; our syntax is much lighter weight than the base syntax 😁. The final product raised a few practical items to work on.

### 1. Excluding blocks from padding
It looks like the first three lines ("A Grammar of Graphics", "By: Andrew Bray", and "A close read of a scatterplot") all get padding instead of showing up like a title block. It looks like the [CSS](https://github.com/qmd-lab/closeread/blob/f2cb5f7ac13e025a1da7c808f534b63d33eaf19e/_extensions/closeread/custom.scss#L23-L30) excludes `<section>` from the padding but not the `<h1>` tags.

Does it make sense to exclude headers from padding and then also provide a class like `.cr-no-padding` or `.cs-join-blocks` that users could append to divs to wrap blocks that they want to spool out without padding? Can that be done directly in CSS (overriding the existing padding rules)?

### 2. Allow for cross-fades to be visible at top of doc
When I move the `.cr-crossfade` trigger up to the Grammar of Graphics header, the plot doesn't show up, presumably because the trigger hasn't met the offset and entrance criteria to get scrollama to change the class to `.cr-active`. What do you think is the best way to solve this?

### 3. Fix Code output vertical alignment
If you look at the first section that triggers both code and output (The Canvas), it looks like the code cell is centered vertically, leaving the plot to extend below the bottom of the screen. It looks like this happens because we [position the sticky thing vertically](https://github.com/qmd-lab/closeread/blob/f2cb5f7ac13e025a1da7c808f534b63d33eaf19e/_extensions/closeread/custom.scss#L37) using `top: 20svh`. Is the solution to wrap sticky blocks in a div container with `display: flex`? I don't really know how flexbox works, I'm largely just inspired by how the [nyt handles this](https://www.nytimes.com/interactive/2022/03/06/books/auden-musee-des-beaux-arts.html).

jimjam-slam:
Sounds like a great test, @andrewpbray! I've been a little distracted with another project in the last few weeks, but I'm hoping I'll be able to start chipping away at these early next month 😊 

I'm also keen to experiment with replacing `.columns` and `.column` with [the Bootstrap CSS grid](https://quarto.org/docs/output-formats/page-layout.html#css-grid). I've used it (actually on this other project I'm working on) and it seems to work really well for responsive layouts, and I think it's probably easier to control where the mobile break happens or to open up other customisation options too.

jimjam-slam:
As far as that second issue goes, we can probably have a boundary condition to ensure the first graphic is always revealed if nothing else!

andrewpbray:
Haha, your ears must have been burning. I have an issue partially drafted starting a discussion on switching from column and columns to lower level grid stuff. I was hitting some roadblocks with columns when trying to tweak, for example, the sidebar to be entirely black a la close read.

Anyhow, I'm hoping to schedule in at least a bit of time each week to think about this and keep the lights on. No worries about responding until you've got some space to work on it!

jimjam-slam:
No worries—I'd like to put at least a bit of time in each week too, so I'll see if I can block it out! (I'm at a conference this week, but from next week I should be able to do some weekly morning meetings again 😊)

PR #21, opened 2024-01-05
Implement gridcss

andrewpbray:


jimjam-slam:
Had a bit of a think after our call about removing that last nested grid for the sticky stack!

I thought on the call that we could just chuck everything directly into the `.cr-layout` and let the grid CSS untangle things using `[data-cr-id] { grid-column: 2 } ` and `:not([data-cr-id]) { grid-column: 1 }`:

```
.cr-layout
  .step
  data-cr-id
  .step
  .step
  data-cr-id
  [...]
```

The problem with that is we know things assigned the same grid cell will stack on the Z-axis, and while we want that for the sticky content, we _don't_ want it for the narrative content. So _that_ content needs to remain inside a `.narrative_col` wrapper:

```
.cr-layout
  .narrative_col
    .step
    .step
    .step
  data-cr-id
  data-cr-id
  [...]
```

The sticky content can be allocated to the second column as before, and the steps can be allocated with `.narrative_col { grid-column: 1 }`.

Then we just have to figure out alignment for the sticky content!

jimjam-slam:
For the horizontal alignment of sticky content, I think just applying a justification should work, as we tried before.

For the vertical, I'm hoping that it might "just work"—but if not, one technique we could try is to actually define our grid as a 2x2, with the top row being of height `1svh` and the bottom being of height `auto`, and then spanning the `.narrative_col` across both rows of the left column while keeping the sticky content on just the first row of the right column.

That way, the second row with the narrative content can grow as needed, there'll be a minimum height for the sticky content and it can be aligned to the centre easily. I could still get the row height and width in JS for scaling ratio calculation—I can get them by checking the styles that define the grid on the `.cr-layout` (or potentially even get the [computed style](https://developer.mozilla.org/en-US/docs/Web/API/Window/getComputedStyle) if needed).

jimjam-slam:
Have just added the changes we tinkered with on the call! At the moment the first `.cr-layout` isn't sized properly—the pink image seems to be getting `.page-full` and `.page-columns` added to it and to all its parents, causing it to be bigger than it ought to be.

jimjam-slam:
Basic poem resizing to fit is working at [47d51f6](https://github.com/qmd-lab/closeread/pull/21/commits/47d51f608f5fc66c215a3c2ecccae5109232ddd8). It rescales when it becomes active, and also when the window resizes.

Only problem is that at some browser zooms (> 100% on my full desktop but > 60% at half height), the element gets pushed down somewhat—I'm not sure why. That can be corrected with an additional transform of `translateY`, but I'm not sure how much to apply—it seems to depend on both content and zoom.

jimjam-slam:
Instead of scaling from the center, you can scale from the top and translate by half the padding. This works well for higher zoom but not for lower zoom. But I'm not sure of a way to detect zoom level reliably, and in any case I'd rather understand what's throwing it off. I might take a closer look at the NY Times's close read!

jimjam-slam:
Okay, the solution in [75cc6bd](https://github.com/qmd-lab/closeread/pull/21/commits/75cc6bde4a2e01f13d4d0e24a359347df682dbaf) works better! It scales _down_ (when the box is too big for the screen) from the top, translating back to the centre to adjust for padding, but scales _up_ (when the box is too small) from the centre. This split strategy seems to work pretty well at a variety of screen sizes and zooms on Firefox, but it needs more testing on Chrome and Safari.

Next up: highlighting parts of it!

I noticed in the NYTimes Close Read that they don't just highlight lines but specific phrases. @andrewpbray, would you prefer to go with the Quarto-like approach we're currently planning, where we specify line numbers to highlight, or would you prefer that users add spans themselves to highlight? (We could also add both approaches, as long as they're not used at the same time...)

andrewpbray:
Hey, this poem looks great! A couple follow ups:

1. Did the pink box resizing thing sort itself out? I couldn't detect any size difference (I did spot an errant `x` that I removed).
2. I tried my darndest to mess up the poem resizing on Chrome and Safari - zooming, resizing - and I couldn't break it!

andrewpbray:
As I've been thinking about the line / phrase highlighting, I've been sniffing around our options for the markdown source that we'll use for these as well as the corresponding Lua types.

Currently, we're using [Backtick Code Blocks](https://pandoc.org/MANUAL.html#extension-backtick_code_blocks) in the markdown and the [Code Block](https://pandoc.org/lua-filters.html#type-codeblock) Lua type. An alternative would be to use [Line Blocks](https://pandoc.org/MANUAL.html#line-blocks) in the markdown and the [Line Block](https://pandoc.org/lua-filters.html#type-lineblock) Lua type.

Here's a little comparison of the source,

````markdown
```
The limerick packs [laughs anatomical]{#cr-laughs}
  In space that is quite economical.
```

| The limerick packs [laughs anatomical]{#cr-laughs}
|   In space that is quite economical.
````

the native AST,

```haskell
[ CodeBlock
      ( "" , [] , [] )
      "The limerick packs [laughs anatomical]{#cr-laughs}\n  In space that is quite economical."
  , LineBlock
      [ [ Str "The"
        , Space
        , Str "limerick"
        , Space
        , Str "packs"
        , Space
        , Span
            ( "cr-laughs" , [] , [] )
            [ Str "laughs" , Space , Str "anatomical" ]
        ]
      , [ Str "\160\160In"
        , Space
        , Str "space"
        , Space
        , Str "that"
        , Space
        , Str "is"
        , Space
        , Str "quite"
        , Space
        , Str "economical."
        ]
      ]
  ]
```

and the HTML,

```html
<pre><code>The limerick packs [laughs anatomical]{#cr-laughs}
  In space that is quite economical.</code></pre>
<p>The limerick packs <span id="cr-laughs">laughs anatomical</span><br>
&nbsp;&nbsp;In space that is quite economical.</p>
```

In the Code Block HTML, looks like the markdown span doesn't get processed, whereas it does in the Line Block.  I haven't tinkered with the Code Block type, but I'd guess that string is just going to get sent through the writer verbatim, so we should be able to write the appropriate html span syntax with the filter.

In terms of the sort of wrapping / resizing work you did, would that all breakdown if the text was in `<p>`? Another potential challenge of the Line Block type is that the Lua reference suggests it can't take attributes? (I don't quite understanding why every type couldn't take attributes).

I was looking at the nyt closeread and they both highlight and zoom to the phrase. Highlighting seems straightforward, and zoom more challenging. Any thoughts on how to do that?

andrewpbray:
An aside: if we stick with Code Blocks, we could also suggest [this syntax](https://pandoc.org/MANUAL.html#extension-fenced_code_attributes) to disambiguate from code cells:

```markdown
~~~
The limerick packs [laughs anatomical]{#cr-laughs}
  In space that is quite economical.
~~~
```

demo-block-types.qmd
```diff
@@ -19,7 +19,7 @@ This is an image with the sticky tag on the div.
 :::
 
 :::{cr-id="pink"}
-![This is a caption](pink-desktop.png)
+x![This is a caption](pink-desktop.png)
```
jimjam-slam, line 22: @andrewpbray This `x` was an accidental change, but it seems to be suppressing the propagation of `.page-columns` and `.page-full`! I'm not sure yet why they're appearing, though...

jimjam-slam:
Mmm, it's tricky! I don't have a strong aesthetic opinion on the line block vs the fenced code block, but I agree that not being able to put attributes on the line block could be a showstopper.

> I haven't tinkered with the Code Block type, but I'd guess that string is just going to get sent through the writer verbatim, so we should be able to write the appropriate html span syntax with the filter.

Makes sense! I can see in Pandoc that the [`bracketed_spans` extension](https://pandoc.org/chunkedhtml-demo/8.18-divs-and-spans.html) processes attributes on code blocks. if it's small Lua extension, maybe we can use that code to reimplement it for the spans? (I feel like we could write it ourselves, but 🤷🏻‍♂️)

jimjam-slam:
> In terms of the sort of wrapping / resizing work you did, would that all breakdown if the text was in <p>?

Nope, should be good! The resizing code just gets the height/width of the code block and of the `.sticky_col_stack`. The contents are pretty opaque to it 🙂

> I was looking at the nyt closeread and they both highlight and zoom to the phrase. Highlighting seems straightforward, and zoom more challenging. Any thoughts on how to do that?

If we have spans with ID inside the code block, I think we can essentially run the resizing code using that instead of the block (although I do also need to know the span's position within the block, not just its size... I think I can use [`element.offsetLeft`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/offsetLeft) and `offsetTop` for that). I think we can ensure the zoomed in image clips to the `.cr-layout` so that things don't get too out of control.

andrewpbray:
### I.

If we go with Code Blocks, the main challenge I see is parsing the markdown span syntax into the appropriate HTML. I just remembered: the Lua in Quarto has no regex engine, only a subset called "[Lua Patterns](https://www.lua.org/pil/20.2.html)". I've used them before and they've been fine, but there's a chance the implementation will have some weaknesses.

If we went with Line Blocks, the main advantage would be that we would need to do limited text processing in the filter since the data comes in more structured. Seems like it'd be straightforward to allow:

```markdown
| {cr-id="mypoem"}
| The limerick packs [laughs anatomical]{#cr-laughs}
|   In space that is quite economical.
```
 
And then use the filter to make a line block (dropping the first line) that's wrapped in a div with whatever attributes get passed through the first line of the line block.

Do you have a strong lean one way or the other - Code Block vs Line Block? I do not.

* * *

### II.

One thing I noticed as I wrote the example above: do we need `.cr-poem`? So far we've used `cr` prepended classes either for the main layout or for a transition trigger. All of our sticky stuff has been identified by the inclusion of the `cr-id` attribute.

It looks like right now that class is used for two things: getting a non-monospace font on there and targeting the element for `rescaleElement` in the js. The latter we'd probably want to do for verbatim code blocks in addition to poem blocks, right? In fact... seems like there are probably lots of sticky elements that we'd like to rescale to fill the viewport.

Would it make sense to have `rescaleElement` look for `.cr-rescale-to-fill` (or `cr-stretch-to-fill`)? Or `rescale-to-fill="true"`? If we wanted the default to be for all Code Blocks (poems and code) to get rescaled, we could use the filter to append that attribute when it finds a sticky object of that type. Alternatively, if we wanted to preserve Code Blocks in their current functionality, that'd be an argument for Line Blocks: the filter could be look for those sticky objects and add classes for `.cr-rescale-to-fill` and `.cr-poem` to get the the non-monospace font.

* **

### III

I realize I'm getting into some scope creep here, but maybe a good time to bring up: for a generic code cell that is evaluated, the output will be both a Pre Code cell and either an Image or another Pre code cell (I think). Right now the default behavior that would seem most natural to me would be for the cell and output to scale normally with the text, as in a normal HTML document, unless it didn't fit vertically or horizontally, in which case I'd expect it to get downscaled.

So are there two scaling functions we want here: one to scale up or down to fill the viewport (`rescaleElement`), and a second one to only scale down if the object extends past the viewport? And would all sticky objects have the latter and only certain ones (either by default or by user pref) have the former?

andrewpbray:
Alrighty, these commits should implement what we talked about today. When Line Blocks appear in the qmd with the first line containing what look like attributes, those attributes (id, classes, attributes), are assigned to a div that wraps the line block and that first line of the line block is removed from the AST.

The example that I added to `demo-block-types.qmd` has all three types of attributes, including a `cr-id`. Miraculously, it just worked: the limerick gets plopped into a sticky section!

jimjam-slam:
I'll fix up the CSS to match when I get a chance on the weekend! It might not actually need any, since a `.line-block` gets the same line break treatment as a `<pre>` but doesn't do a monospaced font 🤔

jimjam-slam:
Also good to note from Pandoc on line blocks:

> Inline formatting (such as emphasis) is allowed in the content, but not block-level formatting (such as block quotes or lists).

jimjam-slam:
Looks Reveal.js's `.r-fit-text`, which does essentially the same rescaling we're doing, it handled by [fitty.js](https://github.com/rikschennink/fitty). I'm taking a look at how that works to see if it does anything we don't!

In the mean time, I'm testing what happens when I remove the checks for `.cr-poem` in the following two places:

 https://github.com/qmd-lab/closeread/blob/443967dbff16da7af1da2fb3887e5e5ad7eece5f/_extensions/closeread/scroller-init.js#L182 

https://github.com/qmd-lab/closeread/blob/443967dbff16da7af1da2fb3887e5e5ad7eece5f/_extensions/closeread/scroller-init.js#L191

For most content types it seems fine, but the graph-viz down the end seems to not scale correctly at zoom > 100%.

I'd be okay with a `cr-rescale="true"` or some such! I will say that everything rescale looks quite nice, but a lot of things look odd when you start looking back at the comparatively tiny narrative column 😂 Maybe there needs to be a max scaling factor?

andrewpbray:
I was just reading back over this PR. It sounded relative to our current work given the name, but it meandered quite a bit into the functionality for the poem. As far as I can tell our branch structure (with their corresponding PRs) looks like:

main < revise-examples < implement-gridcss < feature-poemzoom < feature-poemzoom-columngrid

I'm inclined to try to get as much of this into `main` as possible; I think that'll help me understand the state of the software when I set off to work on a new feature.

Should we consider rolling them in starting with `feature-poemzoom`, then you can switch your `columngrid` PR to be into main directly?

jimjam-slam:
I'm happy to get it into `main`!

PR #22, opened 2024-02-02
Poem zooming

jimjam-slam:
Here's my first swing at the poem zooming effect! I've done some refactoring to keep it as performant as I can (although I suspect this chain-of-transactions pattern I'm naively writing could be done more formally and performantly).

That said, the maths on the zooming isn't working, so you'll notice the poem is zooming off screen right now. Would love a second pair of eyes on this when you have some time, @andrewpbray; you might have a better intuition than me! It's at about L241 of scroller-init.js. 

There's another issue that's a bit of a pin:

Unlike Chrome, Safari appears to masterise text before you apply a transform to it (in some circumstances—I've occasionally managed to get sharp text, mostly when I'm not using a translation, but it's tricky to reproduce!), so small poems that get scaled (scale >> 1) up appear pixelated.

When I look at how Reveal's auto sizing works, they do the opposite: start big and (mostly) scale down. They specify a "natural" width and height and the scaling based off that. So if your natural height and width are high (eg. 1400px), most of the scaling is downward.

So we could specify a large font size for poems and then scale them down.

The trouble is, when a poem is too big for the screen and is scaled down, it still reserves layout space based on its original size. That causes other stickies to miss when it comes to their margins and be pushed down (unless we just decide to apply the transform to _all the stickies?_).


jimjam-slam:
Oops, this was meant to merge into your PR! Closing!

PR #28, opened 2024-06-03
Add ShrinkToFit

andrewpbray:
This PR seeks to add a general `shrinkToFit` behavior that will be applied to any sticky content that overflows its container.

- [x] Add `shrinkToFit.qmd` to test new overflow / shrink functionality.
- [x] Copy `scalePoemFull` as `shrinkToFit` and add to `recalculateActiveSteps`. see commit 48101eb
- [ ] Adapt existing sticky alignment styles to ensure elements are centered in the sticky container.


andrewpbray:
The naive copy-from-scalePoemFull gets part of the way there. The image (which is too wide for the browser window) is shrunk, and the code (which is too tall) is also shrunk.

Screenshot 2024-06-03 at 10 21 37 AM
Screenshot 2024-06-03 at 10 21 48 AM

There are a few issues:

1. When the window is made narrow, there is a point past which the image will no longer resize and so overflows:

Screenshot 2024-06-03 at 10 24 49 AM

2. There appears to be padding at the top of the elements so that tall content still overflows (even though it seems to be rescaled such that the total height is indeed the height of the window).
Screenshot 2024-06-03 at 10 24 27 AM

jimjam-slam:
Safari's rasterisation of scaled up elements is an absolute curse 🫠 See Edge (basically Chrome) vs Safari here:

image

It's happening with the Mermaid diagram, as well as the display math (is that underpinned by SVG?).

jimjam-slam:
Maybe for SVG-based things, there's a way to make them fill the space without transform...

jimjam-slam:
✅ Okay, for the SVG-based diagrams like Mermaid, we shouldn't use `transform` to scale them. Instead, set `with: 100%` on the `.cr-active` wrapper and then override the `max-width` attribute on the `svg` element itself. Because it's set (by Quarto?) as a style attribute, we'll need to use `max-width: 100% !important` to override it (not ideal, but necessary in this case I think). That gets us the smooth scaling we're looking for:

image


For display math, I'm less sure. You can set `width: 100%` and a very height `font-size` (say, `500%`) on the `cr-active` wrapper to get it looking good, but I'm uncertain whether it will cause layout issues too.

jimjam-slam:
Okay, it does.

One interesting thought I got from a solution for [scaling MathML on its own](https://groups.google.com/g/mathjax-users/c/ems8PNage-0/m/DNhpHvANwU4J) is that people are using JavaScript to change the _font size_ on window reload.

Since we're already using JavaScript to change the `transform` scale factor, we're already in JS land... maybe we should do the same and adapt the font size to the window.

I'm kind of already doing this with poems by applying `vw` or `vh` units to the poem font size. Maybe it's worth expanding this approach to math and code blocks.

jimjam-slam:
I've added yet another approach to scaling in #36 — what do you think? I'll copy your code example in and see what I can do there.

jimjam-slam:
I've added your demo to #36 and applied a `font-size: 1.5cqmin` to the code block, which seems to work well. The `2.75` scalar I'm using for the poem is a poor fit, and I think it's going to have to be computed block-by-block based on the content.

The shape of the content is one ingredient in this dynamic sizing, and the size/shape of the window is the other. The CSS `cqmin` and `cqw` units can account for the latter but not the former — that's why we need a scalar based on the former.

But once the code block is sorted out, the image falls into line (it just pokes out because the code block pushes the entire section width out and it takes as much width as it's given).

andrewpbray:
Good work on all this! This sizing stuff is very tricky. Between window size, browser type, block type and shape, there are so many different configurations to consider.

I had a look through #36 . The CSS approach is amazingly simple! So does the magic happen as a result of 

```css
.cr-poem {
  max-height: 100svh;
  max-width: 100svh;
}
```
combined with the use of the `cq-` units for `font-size`?

If indeed that does the trick, should we put these styles either on all `[data-cr-id]` or on `.cr-active`? Could that replace the `shrinkToFit` function? We could then provide a `.no-resize` or something class that users can add to override that behavior.
    
If you hadn't told me, I'm not sure I would have detected the staggering from this method relative to JS transforms. The simplicity of the CSS implementation seems worth it to me. In terms of the constraint related to the shape of the element, the approaches you mentioned: setting the `font-size` via JS or allowing the user to set it both seem like good solutions!

andrewpbray:
Oh, two things that I noticed while fiddling with `shrinkToFit.qmd`...
Screenshot 2024-06-16 at 1 55 24 PM

If you shrink window vertically on the code-output bloc, the code font size shrinks linearly alongside it. Is that caused by this?

```css
        pre.sourceCode, .cell-output pre {
          font-size: 1.5cqmin;
        }
```
While the code shrinks, the figure doesn't change at all. I was able to get the fig to scale alongside it by adding `max-width: 100svh`. Do we want the code to be shrinking ahead of the `max-width` in this scenario?

The other thing that I noticed is that when you first trigger the poems (on the demo blocks doc) it doesn't trigger a zoom-out-from-transparent effect like we had with the transform (and that nytimes closeread has). I thought that was because the `max-height` style is already on the `.cr-poem` so there's no zoom/transition to occur, but shifting it to `.cr-active` didn't seem to change anything. In fact, as I pan up and down over the poem, the transition that I get is a very partial scroll-on and scroll-off the screen (in addition to the opacity). Are you able to replicate that? It not let me know and I can take a short vid.

jimjam-slam:
> Good work on all this! This sizing stuff is very tricky. Between window size, browser type, block type and shape, there are so many different configurations to consider.
> 
> I had a look through #36 . The CSS approach is amazingly simple! So does the magic happen as a result of
> 
> ```css
> .cr-poem {
>   max-height: 100svh;
>   max-width: 100svh;
> }
> ```
> 
> combined with the use of the `cq-` units for `font-size`?

Actually the `max-height` and `max-width` are leftover from the previous approach — they can _probably_ go (they might be good to keep in as a failsafe, although they should probably be `100cqh` and `100cqw` respectively instead). It's really just the `cq-` units on `font-size` making it work!

> If indeed that does the trick, should we put these styles either on all `[data-cr-id]` or on `.cr-active`? Could that replace the `shrinkToFit` function? We could then provide a `.no-resize` or something class that users can add to override that behavior.
> 
> If you hadn't told me, I'm not sure I would have detected the staggering from this method relative to JS transforms. The simplicity of the CSS implementation seems worth it to me. In terms of the constraint related to the shape of the element, the approaches you mentioned: setting the `font-size` via JS or allowing the user to set it both seem like good solutions!

I tend to agree! For the shape constraint, you could have it default to an `"auto"` setting where we try to come up with something based on shape, and then allow the user to override it with an attribute (eg. `cr-size=5`).

> If you shrink window vertically on the code-output bloc, the code font size shrinks linearly alongside it. Is that caused by this?
>
> ```
> pre.sourceCode, .cell-output pre {
>   font-size: 1.5cqmin;
> }
> ```

Yup. `cqmin` uses the shorter of the container width or height; `camas` uses the larger. I applied this to both the source code and the output, but I'm not married to it necessarily being on both! What did you have in mind?

> While the code shrinks, the figure doesn't change at all. I was able to get the fig to scale alongside it by adding max-width: 100svh. Do we want the code to be shrinking ahead of the max-width in this scenario?

Yeah, `font-size` won't cut it for images, only for text (MathML is an interesting case: it's internally rendered by SVG, but you can use `font-size` to control its size). So for raster images, SVG images, videos, etc. we'll need to use `width` and `height` as before (although we hadn't really gotten to them yet). Or `max-width` and `max-height`, if we don't want them growing beyond their native size.

> In fact, as I pan up and down over the poem, the transition that I get is a very partial scroll-on and scroll-off the screen (in addition to the opacity). Are you able to replicate that? It not let me know and I can take a short vid.

Mmm, not sure! I might need to take a closer look. Can you record a quick video?

jimjam-slam:
Oh, sticky blocks with a combination of text and images (eg. source code + figure) might be a challenge to get right! Will have to see how it all ends up looking.

andrewpbray:
[shrink-to-fit.mov.zip](https://github.com/user-attachments/files/15862279/shrink-to-fit.mov.zip)

Here's a quick clip showing 1) the shrink of the text though not the output fig and 2) partial scroll transition of the poem on re-scroll. 2) isn't bad really, I just don't know where its coming from.

Re 1): that does seem a bit tricky; effectively keeping the text / image ratio the same while scaling both with the container size. I'll tinker around to see if I can find a good way to do this.

andrewpbray:
Have been tinkering with `display: flex`, `flex-direction: column` and then `flex-grow: 1` on both the code and the output but haven't been able to get the desired effect yet...

jimjam-slam:
The shrink of the text but not the fig... I hadn't considered this, but the `cq-` units use container width, and since the code is wrapped and the wrapper sits alongside the figure, the figure scales first, and then the code's font size will be based on the remaining space.

I think you're on the right track using flexbox, but determining the natural ratio is hard. This wouldn't be a problem if we were using `vh`/`vw` instead, which uses the _viewport's_ dimensions, but then it won't accommodate layout changes like having a sidebar.

jimjam-slam:
Or maybe we go the other way: wrap the `cqm` units for the `font-size` in `min()` or `minmax()` to protect them from going too small, and put a `flex-shrink` on the figure to allow it to give up space when the code needs it?

PR #29, opened 2024-06-03
initialize crScrollerSection and crScrollerProgress at 0 instead of null

andrewpbray:
Interestingly, this doesn't fully solve the problem. If you try rendering `demo-block-types.qmd`, you'll see that the first transition, to pink-desktop, doesn't trigger adding to the index, so it stays at zero. When you scroll back up to the top, though, it does trigger subtracting from the index, so it goes to -1.

🤔

### DIFF
--- a/_extensions/closeread/scroller-init.js
+++ b/_extensions/closeread/scroller-init.js
@@ -24,8 +24,8 @@ document.addEventListener("DOMContentLoaded", () => {
   const ojsModule = window._ojs?.ojsConnector?.mainModule
   const ojsScrollerSection = ojsModule?.variable();
   const ojsScrollerProgress = ojsModule?.variable();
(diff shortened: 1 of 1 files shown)


jimjam-slam:
I'd need to take a closer look at the problem you're describing, but I've added one other fix in 9131ea6: for reasons I can't explain, I did formatting on the scroller progress number _in the actual progress counter_ instead of in my debugging markdown in the test doc, so it's useless for anything other than display 🤣 That's a number now!

jimjam-slam:
Also, if the first block is high enough up that it's already within view when you start, it might not trigger!

andrewpbray:
I was thinking of that - that the step was already being triggered - but even a very short viewer window (where the step is far below) doesn't solve it. Hmm...

andrewpbray:
I believe this is fixed in #33 .

PR #38, opened 2024-06-28
Fix OJS variables

jimjam-slam:
Changes the names of the Observable JavaScript variables available to users, makes the progress variable numeric (instead of a string), and breaks the direction out as a separate variable.

This cherry picks the commits from #33 so that we can unblock some OJS demos.

### DIFF
--- a/_extensions/closeread/scroller-init.js
+++ b/_extensions/closeread/scroller-init.js
@@ -24,8 +24,10 @@ document.addEventListener("DOMContentLoaded", () => {
   const ojsModule = window._ojs?.ojsConnector?.mainModule
   const ojsScrollerSection = ojsModule?.variable();
   const ojsScrollerProgress = ojsModule?.variable();
+  const ojsScrollerDirection = ojsModule?.variable();
   ojsScrollerSection?.define("crScrollerSection", focusedSticky);
   ojsScrollerProgress?.define("crScrollerProgress", 0);
+  ojsScrollerDirection?.define("crScrollerDirection", null);
   if (ojsModule === undefined) {
     console.error("Warning: Quarto OJS module not found")
   }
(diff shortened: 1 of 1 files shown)


jimjam-slam:
Now that I'm building a demo, I actually see a problem with this: `crScrollerSection` (now `crScrollerName`) previous tracked the index of the _triggering_ block. Now it tracks the name of the _sticky_ block.

For OJS, we're going to want. something to identify the trigger, since we want to drive OJS graphics with the scrolling. I think this needs to be an index rather than an ID, since our triggers may not necessarily have an ID of their own (only stickies do)!

andrewpbray:
It looks like scrollama adds a `data-scrollama-index` attribute to each of the steps. Will that work?

If not, we could add a bit before the scrollama section that gives each of the steps a numbered id (step 1, step 2, etc)?

Actually now that I think about it, the user will have to refer to this id in their ojs code block, right? Seems like we'd want them to make a deliberate choice for which step to use and not have to chase down the appropriate index number.

What are you thinking for syntax? Here's a first cut.

```markdown
:::{focus-on="cr-ojsplot" animate-with="x"}
Let's watch the plot animate.
:::

---{ojs}
\\| label: cr-ojsplot

\\ ojs code that refers to the variable "x"

---

```

jimjam-slam:
We could use that, but iirc Scrollama reports the index of the trigger as part of the callback anyway, so it's probably not necessary 😁

andrewpbray (approved):
Just FYI, in addition to the ojs variable names, it looks like this implements a big of the recalculateActiveSteps refactoring that's found in `rename-steps`.

jimjam-slam:
I think I should probably close this in favour of #40, since I branched that off this PR in order to do the OJS demo :) 

> Just FYI, in addition to the ojs variable names, it looks like this implements a big of the recalculateActiveSteps refactoring that's found in rename-steps.

I think I had to copy it over in order to avoid getting stuck in the middle! Is it out of date now?

andrewpbray:
I... am pretty sure this one is out of date? 

I support deleting 🫡

PR #46, opened 2024-07-23
Add permutation test demo

andrewpbray:
This attempts to add a quarto version of a well-done scrolly that features some very custom js graphics. Right now the html file has been run through pandoc to create md source and first efforts have been made to include some of the js via ojs code cells.

There are a few puzzles that I haven't figured out yet:

- [ ] How to resolve the errors that I'm getting on render related to syntax parsing problems with the ojs cell blocks.
- [ ] Once I shift to using the closeread extension in the source, how we'll want the user to trigger all of the specific functions that update the graphic (more on that below).
- [ ] How to fix the css problems (gray bar on the left side and overriding of the navbar css). This should be straightforward just haven't done it yet!


andrewpbray:
The key functionality that we'd be translating into quarto syntax is the following js:

```js
if (response.index == 0 & response.direction == 'down') {
  transitionZeroDown()
} else if (response.index == 0 & response.direction == 'up') {
  transitionZeroUp()
} else if (response.index == 1 & response.direction == 'down') {
  transitionOneDown()
} else if (response.index == 1 & response.direction == 'up') {
  transitionOneUp()
\\ and so on to transitionEightUp()
```
So he calls a specific function on every combination of step and direction. Would something like the following work?

```markdown
:::{focus-on="alpacas" runjs-if-down="transitionOneDown runjs-if-up="transitionOneUp"}
Text of the step
:::
```

PR #48, opened 2024-07-25
Apply bootswatch theme to narrative col and sticky col

andrewpbray:
One thorny thing that needs to be sorted out: right now the addition of `bg-light` by default screws with the nytimes demo. What's the best way to sort that out so that the nytimes styling takes precedent?

Perhaps by requiring that any secondary theme specify all of the style attributes specified by the bootswatch theme otherwise it's fine to use bootswatch?

### DIFF
--- a/_extensions/closeread/closeread.lua
+++ b/_extensions/closeread/closeread.lua
@@ -2,7 +2,6 @@
 quarto.log.output("===== Closeread Log =====")
 
 -- set defaults
-local debug_mode = false
 local step_selectors = {["focus-on"] = true}
 
 -- Append attributes to any cr line blocks
@@ -86,14 +85,30 @@ end
 
 
 -- Read in YAML options
+
+-- set defaults
+local debug_mode = false
+local narrative_col_bg = "light"
+
 function read_meta(m)
 
+  ----------------
+  -- debug mode --
+  ----------------
   if m["debug-mode"] ~= nil then
     debug_mode = m["debug-mode"]
   end
   
   -- make accessible to scroller-init.js via <meta> tag
   quarto.doc.include_text("in-header", "<meta cr-debug-mode='" .. tostring(debug_mode) .. "'>")
+  
+  
+  -----------------
+  -- tweak theme --
+  -----------------
+  if m["narrative-col-bg"] ~= nil then
+    narrative_col_bg = m["narrative-col-bg"][1].text
+  end
   
 end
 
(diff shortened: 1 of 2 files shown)


jimjam-slam:
If you feel like you've gotten this to a good place as of our call, I'm happy to keep it as is, but I feel like we might benefit from guidance from the Quarto team on this one, since we're essentially guessing at some of the stuff about how SCSSfiles are merged in. Happy to file an issue/discussion if you think it's worth it!

What I do know is that [SCSS files get interleaved](https://quarto.org/docs/output-formats/html-themes.html#theme-options): they have sections like `/*-- scss:defaults --*/` (to set up variables) and `/*-- scss:rules --*/`, and those get individually combined across all SCSS files before the end result is collated (so all the defaults sections stay together, all the rules sections stay together, etc.)

I think ideally our `rules` would just touch layout in the CR sections, and we'd set a few CR-specific variables in the `defaults` that fall back to the existing theming. Then people could set a base Quarto theme, and if they wanted to they could provide additional overrides. For example:

```scss
$cr-narrative-font: $font-family-sans-serif !default
```

What I don't know is:

- The order SCSS files are merged in across extension, file and project
- How the array vs string works (it seems like you have a better idea of this!)

jimjam-slam:
@andrewpbray Also interesting to note that you can provide [format variants](https://github.com/quarto-dev/quarto-cli/discussions/4700#discussioncomment-5246196)... seems like this might be a good way to provide some extra theming out of the box without having to copy those additional styles with a template!

jimjam-slam:
I'll revert https://github.com/qmd-lab/closeread/commit/e1a503828af56aa3c3a9f4bc838f66d0289f0fb7 in `main` in just a sec — I wanted to test whether there was a Quarto bug with `_metadata.yml`. Dang it, GitHub Pages needs branch previews so we don't have to do this in main!

andrewpbray:
Thanks for putting together that discussion post on this. Seeing it all laid out is good validation that...yeah, its kinda complicated!

Good catch about format variants. I hadn't heard about that!

Amen about the branch preview thing. I could set up a netlify workflow for that... 🤔

andrewpbray:
Just for the sake of documentation, this conversation turns a bit on the outcome of convos with the devs: https://github.com/quarto-dev/quarto-cli/issues/10383

andrewpbray:
@jimjam-slam you can now access the PR site at `closeread.netlify.app` (for all the good it does while these style sheets aren't merging!)

PR #60, opened 2024-08-02
Feature presentation mode

andrewpbray:
@jimjam-slam I think this is good to go! Check out the [Guide](https://closeread.netlify.app/guide/#presentation-mode) or the [Reference](https://closeread.netlify.app/reference/#keyboard-hotkeys) for how to use.

Small thing: do you know how to refer to the right and left arrows in the [kbd shortcode](https://quarto.org/docs/authoring/markdown-basics.html#keyboard-shortcuts)?


jimjam-slam:
Ooh, I like the keyboard shortcut! Very reveal-inspired 🥳

jimjam-slam:
I don't know how to do the arrow, I'll afraid :/ Maybe we can have a dig around in the Quarto code to down it out! I'll be on a train this arvo, so I'll try to have a look!

jimjam-slam:
> Small thing: do you know how to refer to the right and left arrows in the [kbd shortcode](https://quarto.org/docs/authoring/markdown-basics.html#keyboard-shortcuts)?

I had a look at the extension, and it honestly looks like it just does platform selection and the just stringifies the shortcut text: https://github.com/quarto-dev/quarto-cli/blob/main/src/resources/extensions/quarto/kbd/kbd.lua

So I reckon we can get by just using the Unicode "[direction]ward arrow" glyphs at U+2190-2193 😊

→ ← ↓ ↑

PR #61, opened 2024-08-04
Feature focus shortcut

andrewpbray:
@jimjam-slam This PR implements a syntactic shortcut to create triggers. It utilizes the Quarto cross-ref syntax. I've spoken with Carlos about it and he's giving it his blessing, or at least doesn't think it'll mess anything up. It is minimally documented [here](https://closeread.netlify.app/guide/#triggers) and there's an example on the Minard map demo.

The Lua filter runs through paragraphs to find citations that look right, removes them from that paragraph, then wraps that paragraph in a trigger Div just as if the user had typed `:::{focus-on...}`.

My next task before this PR is complete is to add another filter that runs on Spans to do the same thing but also bring over other attributes (like `pan-to`). For this the user will use `[@cr-mysticky]{scale-by=2}`.

The parsing *should* be pretty robust since it already relies on pandoc's Cite parser, but we'll learn with use how robust it is. My thought is to not try to get too tricky in allowing shorcuts for muli paragraph blocks or non Para blocks, but guide the user for those to `:::{focus-on}`.

Thoughts?


jimjam-slam:
Oh cool! I like this!

jimjam-slam:
I agree that educating users on using the original syntax in    multi-par trigger situations will be key!

andrewpbray (commented):
looks ok (this was a self-review =))

PR #65, opened 2024-08-05
Revise ojs map

andrewpbray:
Introduce a `crProgressBlock` ojs variable that is tracked a separate scroller and is tied to a `.progress-block` div that can wrap triggers.


andrewpbray (commented):
Looks good (this is a self-review)

andrewpbray:
@jimjam-slam Here is a second way to pass progress to ojs variables. One nice thing about the spanning block approach is that it doesn't seem to suffer from the stuttering that happens is you quickly scroll through the triggers in the second example.

Let me know your thoughts. It feels a bit fragile to move towards wrapping some triggers in an additional Div and not others, but I haven't found any induced bugs so far... 🤞

PR #71, opened 2024-08-09
Add closeread to front page of Guide

andrewpbray:
This is an attempt at putting a closeread section in the guide for a quick here's how to make a closeread doc. It has brought up a few interesting issues that I haven't solved.

- [ ] Users will need to be able to specify alternative article layout classes (besides `column-screen`)in the `cr-section` div. I took a pass at solving this in the filter and I think the filter works but assigning that class within section seems to have some strange behavior (the closeread was initially on the main docs index but it passing it `.column-body` seemed to perturb the non-cr-section content layout. If you have a chance to tinker @jimjam-slam let me know your thoughts on how best to handle this.
- [x] Transitioning from one (verbatim) code block to another as lines get added is visually confusing. We'll need reveal style easing but for now I may just prepopulate the whole doc from the outset then just highlight.
- [ ] This would be a good use case for getting two columns inside the sticky: one for the source and the other for a pic of the output. I tried just tossing in columns but no luck - this'll take some fiddling.


andrewpbray (commented):
looks good, but note that this introduces some to-do comments into the make_cr_section lua function for future treatment of interaction between cr and quarto layout classes (self-review)

andrewpbray:
I haven't figured out the first or third item, but with the cr-section moved into the main `index.qmd`, we don't need to worry about clashes between cr and quarto layouts.

We'll fight that battle another day!

PR #72, opened 2024-08-09
Add devcontainer spec

jimjam-slam:
Very selfish PR here: I only brought my iPad with me on my trip, so I can’t contribute unless I do it through Codespaces 😛

This adds a container spec that includes R, Quarto, a few R packages (mostly just tidyverse and sf, as I was hoping to add a map example as well as a palmer penguins chart example) and other (imo) essentials for VSCode container work, like httpgd (for SVG chart previews).

Unfortunately I can’t get the port forwarding to work on my iPad on the train (it usually works fine though), so I can’t preview my work, but it’s a start!

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,20 @@
+{
+    "name": "Closeread (rocker/r-ver base)",
+    "image": "ghcr.io/rocker-org/devcontainer/r-ver:4.3",
+    "features": {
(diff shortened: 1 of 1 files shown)


jimjam-slam:
I don't think I can make this a draft on my phone, but probably no need to merge this (at least for now)! I just set it up on a branch so I could develop from Codespaces!

andrewpbray:
Hah, very clever! I was wondering what that comment about an iPad was in email...

PR #73, opened 2024-08-09
fix css

andrewpbray:
attempt to constrain width of grid so that big elements don't blow out the right side.


andrewpbray:
@jimjam-slam if you have a moment, could you dip in here and have a look at `index.qmd` at root?

This is just an adapted stickyblocks demo where I removed that x that preceded our pink image that we used as a hack to prevent quarto from throwing it into a Figure environment that applies some classes that blow out the size of the sticky col and shrink the narrative col.
Screenshot 2024-08-09 at 3 56 32 PM
The offending rule may be `width: 100%` here:
Screenshot 2024-08-09 at 3 55 11 PM

But I haven't been able to figure out how to override it, so I'm not sure. What's your diagnosis?

jimjam-slam:
I'm on the train for a good while tomorrow, so I'll jump in! Not 100% sure if I can get browser tools on the iPad, but I'll see what I can do 🤔

jimjam-slam:
Sorry @andrewpbray, it didn't work out - the train is standing room 😮‍💨

jimjam-slam:
Is this PR still relevant @andrewpbray, or should I close it?

PR #76, opened 2024-08-13
Zoom To

andrewpbray:
- [x] Refactor poem scaling code into a separate `zoomTo` function that runs inside `transformSticky`
- [x] Test that it works for: line block line number
- [x] Test that it works for: line block span
- [x] Test that it works for: code cell line number
- [x] Add `hlz` option to do both highlighting and zooming.
- [x] Add `.scale-to-fill` class to apply a transform on an sticky whenever its active


andrewpbray:
@jimjam-slam I believe we have a refactored zooming system that does zooming via `zoom-to`. This allows uses to separately use either `highlight` or `zoom-to` on code or lineblocks. If they want to invoke both, they can use the `hlz` attribute that, on document load, just writes out the analogous `highlight` and `zoom-to` attributes for that trigger.

I'll try to get some docs up and possibly if I have energy before tomorrow, expanding the functionality to display math blocks.

PR #82, opened 2024-08-19
bugfix: figure env messes up grid

### LINKED ISSUE #81: Figure environment messes up css grid
When you add a caption to a sticky image, the `.sticky-col` blows up in width and shrinks the `.narrative-col`. These work:

```md
:::{#cr-trees}
this works: ![a caption](trees.png)
:::
```

```md
:::{#cr-trees}
![](trees.png)
:::
```
This does not:
```md
:::{#cr-trees}
![a caption](trees.png)
:::
```
That is, it appears that a block figure with a caption does not work while an inline figure with a caption and a block figure without a caption do work.

andrewpbray:
- [x] adds a demo file to demonstrate (failed functionality)

Fixes #81


andrewpbray:
We do indeed add the `.column-screen` class!

https://github.com/qmd-lab/closeread/blob/d5583d1c41abd2eb3f8673b403b522c178dbb537/_extensions/closeread/closeread.lua#L77

jimjam-slam:
Are we adding it to the image itself, or just to the section? When I inject logging statements, it seems like we're not adding it to the image, but it's ending up there nonetheless... maybe I need to take another look!

andrewpbray:
This isn't the solution to the problem, but while looking at the Lua I found this at the very bottom of the file

https://github.com/qmd-lab/closeread/blob/d5583d1c41abd2eb3f8673b403b522c178dbb537/_extensions/closeread/closeread.lua#L519

As far as I can tell, I don't think that's actually a function that we define?

I'm just recording that here for now.

_extensions/closeread/closeread.lua
```diff
@@ -71,22 +71,28 @@ function make_section_layout(div)
...
     
-    -- todo: identify quarto layout to use in section
-    --local quarto_layouts = {"column-body", "column-outset", "column-page", 
-     -- "column-page-inset", "column-screen-inset", "column-margin"}
-    local quarto_layout = "column-screen" -- default
-    --if list_includes_any(div.classes, quarto_layouts) then
-    --  quarto_layout = ""
-    --end
+    -- get existing style on .cr-section so we don't lose user styles
+    local existing_style_str = div.attributes.style or ""
```
andrewpbray, line 75: This looks like it's doing everything with inline styles - assuming users are using that approach and then using that appraoch to write inline styles here in the filter. What's the reason for that over classes? Is this so that we have a lock on specificity and will be able to override class-styling with inline styling?

docs/gallery/demos/sticky-blocks/index.qmd
```diff
@@ -0,0 +1,54 @@
...
+Just about any type of element that can be included in a Quarto doc can be flagged as a sticky. This demo shows many examples.
+
+The current fix allows `.cr-section` to work, provided it's a "first level" child of the `main` content. For example:
+
+::::{.cr-section}
+
+This is an image that includes a caption. @cr-inline-working
+
+:::{#cr-inline-working}
+This works: ![a caption](trees.png)
```
andrewpbray, line 19: heyo!

docs/gallery/demos/sticky-blocks/index.qmd
```diff
@@ -0,0 +1,54 @@
...
+![](trees.png)
+:::
+
+::::
+
+But things fall apart if you put one inside a `section` (which Quarto creates from an `h2`).
+
+## Images
+
+::::{.cr-section}
```
andrewpbray, line 34: Oh, this is unexpected behavior but to me but very interesting.

_extensions/closeread/closeread.lua
```diff
@@ -71,22 +71,28 @@ function make_section_layout(div)
...
+
+    -- NOTE - we can't use quarto's .column-* classes, so we have to add those
+    -- styles manually (split them on `;`, add new ones, concat again)
+    local existing_style_table = {}
+    for str in string.gmatch(existing_style_str, "([^;]+)") do
+      table.insert(existing_style_table, str)
+    end
+    table.insert(existing_style_table, "grid-column: screen-start / screen-end")
```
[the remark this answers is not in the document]
andrewpbray: and does adding `.page-full.page-columns` on the parent element set these to be the far left and far right grid lines, whereas without those classes they'd be elsewhere (we see inside the h2 section on sticky blocks that they're grid lines in the middle col)?

_extensions/closeread/closeread.lua
jimjam-slam: I'm ashamed to say that I digit we could just write our own class that mimics `.column-screen` 😂 Should probably revert this and just change the class name!

docs/gallery/demos/sticky-blocks/index.qmd
jimjam-slam: Yeah! This was the thing that helped be understand why `.page-full.page-columns` is propagated: only _direct_ grid children can be positioned.

_extensions/closeread/closeread.lua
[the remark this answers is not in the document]
jimjam-slam: Yes! Either `.page-full` or `.page-columns` (I can't remember which) sets the positions and names of the Quarto grid tracks; ` screen-*` are the outer ones. Adding those classes to children in the body allows the grid to reach _their_ children, which is what allows you to set a column class on an element in your document even when it's nested inside other stuff.

_extensions/closeread/closeread.lua
andrewpbray: makes sense to me! would make the layouts far more transparent to users (and tweakable).

_extensions/closeread/closeread.lua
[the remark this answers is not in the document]
andrewpbray: Ok, that makes sense. The direct-child thing aside, this is a nice way to refer to the columns that we'd like to span!

jimjam-slam:
With https://github.com/qmd-lab/closeread/pull/82/commits/f1eb9b24b1a197a5dc08a2d08fc24fd6e3555f6e, this is ready for review to merge in @andrewpbray! The change to propagate classes back is based on the [similar Quarto code](https://github.com/quarto-dev/quarto-cli/blob/d85467627aae71c96e3d1e9718a3b47289329cde/src/format/html/format-html-bootstrap.ts#L1163)

_extensions/closeread/closeread.js
```diff
@@ -28,14 +28,23 @@ document.addEventListener("DOMContentLoaded", () => {
...
+    https://github.com/quarto-dev/quarto-cli/blob/
+      d85467627aae71c96e3d1e9718a3b47289329cde/src/format/html/
+      format-html-bootstrap.ts#L1163C1-L1186C7 */
+  const ensureInGrid = el => {
+    const parent = el.parentElement
+    parent.classList.add("page-columns", "page-full")
+    if (isDocumentMain(parent)) {
+      return
+    } else {
+      ensureInGrid(parent)
```
andrewpbray, line 42: Is it necessary to give these classes to the whole ancestry of `el` vs just the direct parent?

_extensions/closeread/closeread.js
```diff
@@ -28,14 +28,23 @@ document.addEventListener("DOMContentLoaded", () => {
...
-    console.error ("No `main#quarto-document-content` found. Closeread is not functioning properly.")
-  }
-  docMain.classList.add("page-columns", "page-full")
+  // attach layout classes to parents of `.cr-section`s up to main.content
+  /* this replicates quarto <= 1.6 functionality:
+    https://github.com/quarto-dev/quarto-cli/blob/
+      d85467627aae71c96e3d1e9718a3b47289329cde/src/format/html/
+      format-html-bootstrap.ts#L1163C1-L1186C7 */
+  const ensureInGrid = el => {
+    const parent = el.parentElement
```
andrewpbray, line 37: Boy, these DOM selectors are pretty great! Don't think there's such a thing in lua / pandoc lua API.

andrewpbray:
Oh hey: since this was originally into `main`, looks like we get that action:

[https://closeread-pr.netlify.app/gallery/demos/sticky-blocks/](https://closeread-pr.netlify.app/gallery/demos/sticky-blocks/
)

Looks like it's not fixed there...hmm... what's going on here.

Sidenote: `sticky_blocks.qmd` showing up makes me think: do we want to showcase all of these tests as "Demos" or is there some tension there that would be better resolved using Demos for clear functionality vs some other form of less public test doc?

_extensions/closeread/closeread.js
jimjam-slam: Oh, ofc! I completely forgot that we discussed that. Let me check!

_extensions/closeread/closeread.js
jimjam-slam: Yeah, the DOM API is looooovely

jimjam-slam:
> Oh hey: since this was originally into `main`, looks like we get that action:
> 
> https://closeread-pr.netlify.app/gallery/demos/sticky-blocks/
> 
> Looks like it's not fixed there...hmm... what's going on here.

Are our actions deploying on push to a branch or on opening a PR into a branch? I'm not sure exactly when `pull_request` actions fire with respect to a PR! We probably also don't want every PR opened into `main` overwriting our published production docs either...

> Sidenote: `sticky_blocks.qmd` showing up makes me think: do we want to showcase all of these tests as "Demos" or is there some tension there that would be better resolved using Demos for clear functionality vs some other form of less public test doc?

Mmm, I do wonder if we'd be better cordoning off tests. They could potentially stay on the site if it's easiest to keep them there (to take advantage of our current CI/CD), but maybe putting them in a separate category and labelling them as being for internal testing purposes is a good idea (once the work is fully merged in).

_extensions/closeread/closeread.js
jimjam-slam: Looks like it still works without those four lines! However, there's another problem — will address in the main thread.

jimjam-slam:
Have stopped going all the way up the DOM in https://github.com/qmd-lab/closeread/pull/82/commits/c9c6a9b673d6e3144c5ca6b4244e45b6144fb197. Returning to the sticky blocks demo, we do still have a problem. Our style stretching it across the screen is being overwritten by a more specific one from Quarto:

```
.page-columns.column-page-left .page-columns.page-full>*,        // this is the one that's triggering
.page-columns.column-page-left>* {
  grid-column: page-start/body-content-end;
}
```

I think the easiest way around this is probably just to set our CSS grid style as `!important`, which should override just about anything Quarto sets regardless of the specific page layout Quarto is using. This will change as we decide how to accommodate sidebars, but I think it makes to evaluate whether to change this style as part of that work rather than here.

jimjam-slam:
(When we merge this one, we should probably squash it! There's been a lot of back-and-forth in the commits 😅)

andrewpbray:
It feels like this is one of those rare cases when `!important` really is called for!

In terms of whether sticky blocks should be a demo or a test, I agree with your thoughts there. If we don't cordon off tests, among other things, it will bloat the docs with stuff that users might be kinda like wtf on. How about this approach for testing (just spitballing...):

1. Put any test qmd files in `docs/tests`
2. Add `project.render["*.qmd", "!tests"]` to `_quarto.yml`
3. Create a new `_quarto-tests.yml` file with `project.render["tests"]`
4. In both actions, add an additional step before `quarto publish` where we call `quarto render --profile tests`.

When the profile metadata gets merged in, hopefully the "tests" glob overwrites the "!tests" one?

andrewpbray:
🎉 🎉 🎉

PR #98, opened 2024-10-15
Ensure predictable layouts for stickies with code output

andrewpbray:


andrewpbray:
I suspected that the fixes from #81 will at least fix the third example that shows up in `docs/gallery/demos/code-output/index.qmd` - that of a plot produced by ggplot2 code. But when I tried to merge it into this branch, it didn't fix it and indeed `sticky-blocks` looked a bit different. Oooh, maybe this is a prime use-case for a dev container...

PR #106, opened 2024-10-20
Changing browser zoom breaks Scrollama progress updates

### LINKED ISSUE #101: Changing browser zoom breaks Scrollama progress updates
On https://closeread.netlify.app/gallery/demos/ojs-variables, scroll part way through the OJS animation, then change browser zoom and continue to scroll. The Scrollama progress variables no longer update.

Perhaps Scrollama needs to be rebooted or some such when browser zoom changes?

Observed on Safari and Chrome (and I've just realised that the OJS demo fails entirely on Chrome! 🪦)

jimjam-slam:
Fixes #101. Based on https://github.com/russellsamora/scrollama/issues/145#issuecomment-1764582014


jimjam-slam:
Deleting because I've accidentally branched off main and don't want to cause problems!

PR #107, opened 2024-10-20
Fix Scrollama problems on browser resize/zoom

### LINKED ISSUE #101: Changing browser zoom breaks Scrollama progress updates

jimjam-slam:
Fixes #101. Based on https://github.com/russellsamora/scrollama/issues/145#issuecomment-1764582014


jimjam-slam:
Note that I thought I would need to manually reattach the Scrollama event handlers, so I moved them into named functions. I ended up just needing to call `.resize()` methods on the scroller objects, so the refactor into named functions ended up not being necessary. I think it improves readability and is worth keeping, but I can undo that part if you disagree!

docs/index.qmd
```diff
@@ -40,7 +40,7 @@ Step three: flag an element to become a sticky. [@cr-doc]{highlight="14-18"}
 Step four: create a trigger to make the sticky appear. [@cr-doc]{highlight="12"}
 :::
 
-:::{#cr-doc filename="myfirstcr.qmd"}
+:::{#cr-doc filename="myfirstcr.qmd" .scale-to-fill}
```
andrewpbray, line 43: The drama! :)

andrewpbray:
Testing it locally on chrome, the freeze-on-resize issue was still there (as was some jumpiness).

https://github.com/user-attachments/assets/cef46957-2c08-4394-8f95-07ff20f2f49c

I wanted to test it on the preview site, so I reran that job and, presto, no jumpiness and the resize/reload issue is fixed! Good to know that serving that stuff locally won't behave in quite the same manner.

Alright, I'll proceed with the review!

andrewpbray:
Small note: in [bcfcc66](https://github.com/qmd-lab/closeread/pull/107/commits/bcfcc66ab4c0972b3180aec0c2029b97ae0d4cd3) I moved the second hotkey event listener inside the on-document-load after looking into it a bit a deciding that that probably makes the most sense for all of our listeners.

andrewpbray (approved):
Looks good!

_extensions/closeread/closeread.js
```diff
@@ -2,6 +2,8 @@
 // closeread.js //
 //==============//
 
+// import throttleDebounce from "https://cdn.jsdelivr.net/npm/throttle-debounce@5.0.2/+esm"
+
```
jimjam-slam, line 6: I'd originally intended to use this package, but I didn't really want t bite off a bundling toolchain (although we could probably load it via a script tag in the HTML, as we do our existing dependencies). I ended up using the native browser `setTimeout()` instead, so this can go (although I could try loading it via script and using debounce, which would prevent `.resize()` being called multiple times). Not sure there's any noticeable performance impact currently, though.

andrewpbray (approved):

PR #109, opened 2024-10-23
Enhance docs

jimjam-slam:
Have filled the docs out a bit more and done some light editing! Still a few things missing, but I think they're mostly connected to fixes and features we haven't implemented yet. What do you think, @andrewpbray?


andrewpbray:
This is excellent! As I look through your work, if I have small edits, I'll make them directly, but please feel free to object if you feel strongly!

docs/index.qmd
```diff
@@ -28,18 +31,20 @@ format: closeread-html
 
 ::::{.cr-section layout="overlay-center" style="font-size: 1.5em; background-color: rgba(39, 128, 227, 0.2);"}
```
andrewpbray, line 32: I tried setting the opacity to 1 here but I couldn't get it to override the `sourceCode` opacity that Quarto uses.

jimjam-slam: Mmmm, this was the background of the code block, right? Do we have that set in our theme already?

andrewpbray:
Before I implement this whole hog, I have a hot take to run by you: I'm not a fan of callout boxes 😬 

I find my eye reads them first (because of the box, because of the color), before i get oriented to the hierarchy of the sections. Indeed they call my attention, but usually in a way that ruins my flow of parsing a page in a more linear fashion.

If you're strongly pro-callout, I'm totally fine keeping em in, but just thought I'd propose an edit :)

jimjam-slam:
Haha, no worries at all — I do like callouts generally, but I don't have strong opinions on them here.

I tend to also use them for parenthetical stuff, too, and that stuff (ie. anything marked`.callout-note`) could probably be chucked into the margin or under a footnote instead!

docs/guide/styling.qmd
```diff
@@ -0,0 +1,33 @@
+---
+title: Styling
```
andrewpbray, line 2: This is a great addition!

andrewpbray (approved):
This looks great!

andrewpbray:
Am I good to merge this one?

docs/index.qmd
jimjam-slam: Because of the heavy nesting of code blocks, you need to actually do it from a stylesheet rather than inline.

```scss
.cr-section {
  .sticky-col {
    .sticky-col-stack {
      div.sourceCode {
        background-color: rgba(233, 236, 239, 1);
      }
    }
  }
}
```

I'm a bit worried this might override user theming (I don't see it set as a Bootstrap var, but it might come through Quarto from a syntax processor), but let's see how we go. Have added this now :)

PR #110, opened 2024-10-25
Add plot build-up demo

jimjam-slam:
This adds a demo to the gallery where we build up a ggplot one detail at a time, allowing us to discuss a plot with two potential messages. There's some light customisation too (I've added it as a raw HTML block rather than linking to a stylesheet), but nothing wild.

One big thing to note is that it only works because we're not using `.scale-to-fill` — if you turn that on, each new step scales up from its natural size when it become visible. That's not a problem for most R/Python/Julia plots, which will be rasterised by default (and they'll still scale down, so as long as you make them big enough for desktop, it's fine).

We could fix that issue by running `scaleToFill()` on _every_ sticky on page load and on window resize — that way they scaled correctly before you get to them. Not sure if it would interfere with the focus effects, though; maybe you'd want to be more careful on the window resize in terms of checking where you are. What do you think, @andrewpbray?

### DIFF
--- a/docs/gallery/demos/build-up-plots/index.qmd
+++ b/docs/gallery/demos/build-up-plots/index.qmd
@@ -1,0 +1,191 @@
+---
+title: Build up plots
+description: "If your plot is too complicated, try introducing elements one at a time."
+format:
+  closeread-html:
+    code-fold: true
+    cr-section:
+      layout: overlay-left
+knitr:
+  opts_chunk: 
+    dev.args:
+      bg: transparent
+---
+
+In this demo, we'll recreate a plot from the documentation of Allison Horst's `{palmerpenguins}` R package, which provides access to penguin data from Dr. Kristen Gorman and the Palmer Station in Antarctica.
+
+```{r}
+#| label: setup
+#| echo: true
+#| output: false
+library(dplyr)
+library(ggplot2)
+library(palmerpenguins)
+
+penguins |>
+  filter(
+    !is.na(species),
+    !is.na(flipper_length_mm),
+    !is.na(body_mass_g)) ->
+penguins_forplot
+```
+
+The plot looks like this:
+
+```{r}
+#| label: original-plot
+#| echo: true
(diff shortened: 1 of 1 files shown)


andrewpbray:
Ooo, this is lovely. I few thoughts:

1. The issue with `.scale-to-fill` is one that I figured we'd bump into at some point - guess that point is now!  I think the behavior that you mention - doing the scaling on load - is the one that makes more sense. The zoom-in effect is kinda cool, though! I guess you could get the same effect by using `scale-by` when you trigger a new sticky. If combined with `.scale-to-fill` and the user wants a zoom in effect of 25%, we'd probably want to first scale-to-fill, then scale-by .80 (all before cr-active is added), then add scale-by 1.25 to achieve an overall effect of, "scale the sticky by 25% so that it fits the viewport"

2. Was it intention to not propagate the annotated axis arrows in the second plot?

3. I like the blue background. Thoughts on appropriating that for the index?

jimjam-slam:
> Ooo, this is lovely. I few thoughts:
> 
> 1. The issue with `.scale-to-fill` is one that I figured we'd bump into at some point - guess that point is now!  I think the behavior that you mention - doing the scaling on load - is the one that makes more sense. The zoom-in effect is kinda cool, though! I guess you could get the same effect by using `scale-by` when you trigger a new sticky. If combined with `.scale-to-fill` and the user wants a zoom in effect of 25%, we'd probably want to first scale-to-fill, then scale-by .80 (all before cr-active is added), then add scale-by 1.25 to achieve an overall effect of, "scale the sticky by 25% so that it fits the viewport"

Reckon I might split this off to a separate issue to pick up during/after the contest!

> 2. Was it intention to not propagate the annotated axis arrows in the second plot?

It wasn't, but you could also choose to leave them there!

> 3. I like the blue background. Thoughts on appropriating that for the index?

I'm happy for it to be used if you like it, but I also quite like the bright blue tbh!

PR #112, opened 2024-10-25
Move cosmo theme to whole gallery (not just demos)

### LINKED ISSUE #111: themeing is broken on closeread.dev
The pages in the gallery (which run their own css) have a different nav bar than the main page:

cosmo

nyt

Looks like the `cosmo` theme isn't percolating down to these pages (the blue navbar is the default I believe).

Maybe we didn't run `sass` before merging? That'd be a good thing to put on CI I guess? (not sure if `sass` is on the image)

jimjam-slam:
Fixes #111


andrewpbray (approved):
hey, its nice when a bug has an easy fix :)

PR #113, opened 2024-10-25
YAML style options

jimjam-slam:
Proof of concept here in 8128fb9: if the Lua finds a property defined in the YAML under `format.closeread-html.cr-section.style`, it writes an HTML style block that creates a CSS variable under this name (prefixed with `--cr-`), targeting the whole document.

In our SASS, there is a default definition for this var covering the whole doc too, and the corresponding properties in the SASS point to it.

This way:

1. if a user chooses to manually override the original CSS property (with enough specificity), a default or user-supplied CSS variable value is ignored
2. if a user overrides the CSS variable value for the whole doc in their own CSS stylesheet, that ought to take precedence over ours (I think)
3. if a user overrides the CSS variable value for a section, that should take precedence in that section
4. if a user supplies a YAML option, that should override everything (unless they've overridden the original property, see 1).

I think this makes sense as an appropriate hierarchy of customisation: we can supply a sensible default, provide easy customisation paths for those who want them, and stay out of power users' way.

This needs to be expanded t more properties, and we might want to choose defaults that lean on Bootstrap if it's available (although I'm not sure we can do that from our SASS, at least until Quarto 1.7).


jimjam-slam:
Have added some more CSS variables:

```scss
:root {
  --cr-trigger-background-color: var(--bs-primary);
  --cr-trigger-text-color: var(--bs-light);
  --cr-trigger-border-radius: 1em;
  --cr-trigger-max-width: 527.98px;
  --cr-trigger-outer-margin: 5%;
  --cr-sidebar-background-color: var(--bs-primary);
  --cr-section-background-color: var(--bs-secondary-bg);
  --cr-font-family: var(--bs-body-font-family);
  --cr-font-size: 1.15rem;
}
```

These are all hooked up in the YAML (no `--cr-` prefix used there), but not all are hooked up to the corresponding CSS properties (mostly because I wanted to wait until the PR that tinkers with layout and padding is merged).

I've tied some, but not all, of these to Bootstrap CSS variables. Note that the if the corresponding Bootstrap variable is not available, the `var()` call falls through to the user agent default rather than reverting to some previously defined value. For us that shouldn't be a problem because:

(a) those variables will only be unavailable if `theme: none` is set
(b) Quarto provides very little CSS if `theme: none` is set (ie. the main font defers to the user agent stylesheet) anyway
(c) If a user provides values for our CSS variables and scopes it to `.cr-section` instead of `:root`, that should take precedence regardless of whether they do it in a stylesheet, in an inline code block or directly on the element.
(d) as long as the no-theme fallback doesn't actively _break_ functionality, I'm fine with it being ugly — `theme: none` is _supposed_ to be minimal style-wise

Happy to tweak these defaults or names, though!

jimjam-slam:
I've caught this up to `dev` to make the merging easier. I've hooked up the rest of the proposed options, and I've altered the existing gallery docs and home page to use those options. (In the case of NYTimes and Minard, I've trimmed down the CSS where I can, but I wasn't sure how much oft the Minard CSS is actively being used, so I didn't go too wild).

A few things that cropped up while I was working:

1. Do we actually apply .cr-poem to poems? It looks like we don't anymore — the highlight code appears to function when it finds `.line-block` inside `.sticky`, so I've made the poem font function the same way.
2. I'm having a lot of trouble passing through comma-separated lists of fonts through — when they go through the Lua, everything after the first comma is lost. Not sure why this is, but I'm going to do some Lua debugging to try and pinpoint it.

jimjam-slam:
Fixed the issue with strings that contains commas (primarily font lists)! This is ready for review I think, @andrewpbray!

andrewpbray:
Taking a look now!

docs/gallery/examples/auden-poem/index.qmd
```diff
@@ -4,11 +4,38 @@ image: gabbert.png
 author: Elisa Gabbert
 format:
   closeread-html:
-    css: nytimes.css
+    # css: nytimes.css
     remove-header-space: true
     debug-mode: false
+    cr-section:
```
andrewpbray, line 10: Oh, I love this! What a great example of how much you can do through these yaml options!

I think this is spot on with the exception of two things: font size is 1.25rem on NYT and there's also a min-width on the narrative col that keeps it from expanding when you widen the screen. After I finish this review I'll go ahead and push those changes.

andrewpbray: A follow-up: this is actually an improvement over the existing version. If you look at the mobile version, we were using two slightly different blacks and it looks just slightly off. Here it looks great!

andrewpbray (commented):
This is awesome!

Two main discussion points:

1. After seeing how you disarmed the bootswatch theme in the Demos, I'm inclined to either switch from `$primary` to `$dark` or just remove those default links to the bs variables.
2. It might make sense to rename at least some of the `trigger-` options `narrative` options to reflect what they're doing in the css.
3. I'm kinda inclined to remove `cr-section` under `closeread-html` in the yaml entirely. The original plan was that under `closeread-html` there would be `cr-section`, `cr-trigger`, and `cr-sticky`, where you could pass global options to those three core components. As this is maturing though, that doesn't seem like the most useful factoring. Seems like we could just use `cr-style` to differentiate that this isn't a generic html key.

jimjam-slam:
Awesome! Today's a bit hectic, so it might take me some time to respond to things individually, but the tl;dr is I'm happy with all your comments and proposed changes :) Will see if I can get some time tonight or early tomorrow morning to make your suggested changes and document them over on #109 :)

_extensions/closeread/closeread.scss
```diff
@@ -162,12 +169,12 @@
     grid-column: 1;
     z-index: 1;
     transition: opacity 0.5s ease-in-out;
-    max-width: 527.98px;
+    max-width: var(--cr-trigger-max-width);
```
andrewpbray, line 172: If this is only being applied in the overlay layouts, should it be called something like `overlay-narr-max-width`?

_extensions/closeread/closeread.scss
```diff
@@ -135,12 +141,13 @@
       grid-column: 1;
       z-index: 1;
       margin-inline: auto;
+      background-color: transparent;
       
       .narrative {
-        background-color: rgba(17, 17, 17, .85);
-        color: white;
+        background-color: var(--cr-trigger-background-color-overlay);
```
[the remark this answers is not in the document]
jimjam-slam: Given this actually is the individual narrative block, I would argue that trigger makes more sense... although I suppose it also covers narrative content that isn't actually a trigger. Given my comment above, I might have oversteered in preferring trigger to narrative in the docs! Will try to adjust back in #109

_extensions/closeread/closeread.scss
jimjam-slam: Yup, sounds good!

andrewpbray:
These last two commits are a first take at a `sidebar-width` option. It occurred to me that there might be two options here: fixing the width of the sidebar, as I've done, or chance the col allocation of in the grid to allow that width to change with screen width.

Should we have both? If so, strong opinions on names?

_extensions/closeread/closeread.scss
```diff
@@ -211,6 +211,7 @@
     grid-column: 1;
     margin-inline: 0;
     background-color: var(--cr-sidebar-background-color);
+    width: var(--cr-sidebar-width);
```
jimjam-slam, line 214: Instead of targeting the `width` of the `.narrative-col` with the variable, I would probably target the `grid-template-columns` of the `.sidebar-left`, so `1fr 2fr` becomes `var(--cr-sidebar-width) 2fr`, and `--cr-sidebar-width` defaults to `1fr` (unless you think it should be fixed size by default).

The `fr` units allocate space left over after allocating fixed space to other columns, so if column 1 (the sidebar) gets `400px`, for example, the second column gets everything leftover regardless of how many `fr` it uses (they're ratio units).

Same deal for `.sidebar-right`, just with the column order flipped I think!

andrewpbray: Yep, makes a lot of sense, I'll get that changed.

docs/gallery/examples/auden-poem/index.qmd
jimjam-slam: I've adjusted the sidebar width in the YAML from `400px` to `minmax(250px, 1fr)`. That way, it goes 1:2 with the sticky content, but only down to a minimum of 250px. It's pretty tough to read both the poem and the narrative at that particular width, though, and I don't have a good way to redefine the mobile breakpoint via CSS variables (they can only be values for other properties, not things like the parameters for media queries). Maybe there's a JS way to do it, but I'm not sure. That might have to be a post-contest goal.

_extensions/closeread/closeread.scss
jimjam-slam: Have done both for the NYT demo!

_extensions/closeread/closeread.scss
```diff
@@ -351,4 +306,21 @@ body.cr-removeheaderspace {
...
+  --cr-trigger-background-color-overlay: var(--bs-primary);
+  --cr-trigger-text-color-overlay: var(--bs-light);
+  --cr-trigger-text-color-sidebar: var(--bs-dark);
+  --cr-trigger-border-radius: 1em;
+  --cr-trigger-max-width: 527.98px;
+  --cr-trigger-outer-margin: 5%;
+  --cr-trigger-font-family: var(--bs-body-font-family);
+  --cr-trigger-font-size: 1.15rem;
+  --cr-poem-font-family: Iowan Old Style, Apple Garamond, Baskerville, Times New Roman, Droid Serif, Times, Source Serif Pro, serif, Apple Color Emoji, Segoe UI Emoji, Segoe UI Symbol;
+  --cr-sidebar-background-color: var(--bs-primary);
```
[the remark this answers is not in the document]
jimjam-slam: Have removed the BS var refs and gone for static default colours! We could adjust them to taste more, but I'm inclined to settle on something and leave further tweaks for the contest, particularly as they're easy to tweak. Left the narrative font family to match Bootstrap.

PR #123, opened 2024-11-01
Fix cr-sections inside sections (eg. when h2 is used)

jimjam-slam:
Reverts commits from #82 only sending `.page-columns.page-full` up to direct parents — now they go all the way back up to `main`. Fixes the width I the Minard demo, where the `.cr-section` is inside a `section` due to the use of an `h2`.

Also bumps version to 1.0.1!


jimjam-slam:
Have also incorporated the sidebar-right fix suggested by @lattaai1 (but done properly on the SCSS).

andrewpbray (approved):
Looks good!

PR #127, opened 2024-11-03
v1.0.1: sidebar-right fix, section fix, CR Prize announcement

jimjam-slam:


andrewpbray (approved):
looks good!

andrewpbray:
One thing: I didn't see a map file change there for the scss file, but I also don't see it in .gitignore. Unless browsers need it, I'd be inclined to remove it altogther since the repo will have the css file used by the extension as well as the scss file that serves as its source.

andrewpbray:
Reader to merge I think!

PR #128, opened 2024-11-03
feat: new focusing effect: crop figures reversibly 

zzzhehao:
When presenting complex figures, simple pan or zoom effects may not optimally guide the audience's attention. To address this, I've added a crop effect to remove distracting parts of the figure. Use [@cr-myfigure]{data-crop="20%, 20%, 20%, 20%"} to specify the cropping portions on all four sides of the figure (top, right, bottom, left). To remove all existing cropping effects, simply call the figure without any specification: @cr-myfigure.

Brief introduction has been added in `docs/guide/focus-effects.qmd`. 

Here [a quick demo video](https://github.com/zzzhehao/closeread/blob/feat-crop-demo/docs/assets/closeread-crop-demo.mov).

### DIFF
--- a/_extensions/closeread/closeread.css
+++ b/_extensions/closeread/closeread.css
@@ -38,12 +38,20 @@
 .cr-section .sticky-col {
   grid-row: 1;
 }
+.cr-section .sticky-col .sticky {
+    transition: transform 3s ease-in-out;
+}
+.cr-section .sticky-col .sticky img {
+    transition: clip-path 1s cubic-bezier(0.4, 0, 0.2, 1);
+}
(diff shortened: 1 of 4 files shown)


andrewpbray:
Thanks for the PR!

We're very keen to work on this with you but it'll be about a week before we can take a look.

zzzhehao:
> Thanks for the PR!
> 
> We're very keen to work on this with you but it'll be about a week before we can take a look.

No stress, I also opened PR completely unannounced. I got the idea and did some tiny modification for my own work, I just pushed to share it but it most probably needs some polishing from the developing side.

PR #173, opened 2025-02-25
Post-prize updates

jimjam-slam:
- [x] Update site banner (co-authored by @KittJonathan)
- [x] Add entries to gallery


jimjam-slam:
@andrewpbray I've updated the gallery page here to include the winners of the Closeread Prize. Does that look good to you?

andrewpbray:
Hey, this looks great - nice work with the ejs templating! Just one small quibble: the wide of the text column under the Closeread Prize heading is very wide and makes it somewhat difficult to read. Actually, the same goes for our feature demo table. What do you think about keeping both of those sections narrower and then bumping it out for the sections with cards?

andrewpbray (approved):

andrewpbray:
Another thing that comes to mind: we may want to email the folks whose content we link to to let them know we're doing so (so that they hopefully keep those links fresh).

jimjam-slam:
> Just one small quibble: the wide of the text column under the Closeread Prize heading is very wide and makes it somewhat difficult to read. Actually, the same goes for our feature demo table. What do you think about keeping both of those sections narrower and then bumping it out for the sections with cards?

We can do that! It's currently at the standard Quarto body width, so we'd have to set a lower max-width on the text and the table (and either keep left justified or centre). Here's how it looks with 800px max width and entering — what do you think?

image


> Another thing that comes to mind: we may want to email the folks whose content we link to to let them know we're doing so (so that they hopefully keep those links fresh).

Sounds good to me! Maybe Curtis can send an email out to all the entrants, since I think he's the one with access?

andrewpbray:
Hmm, that looks a bit odd to me with the headers back on the original width. If that's not easy to change, though, I'm a-ok with just going for the default for now. We can always come back and revise these pages at a later date if we have the inspiration.

jimjam-slam:
No worries, I'll und that and merge it in!

PR #181, opened 2025-07-12
Widget sizing work

jimjam-slam:
Still WIP!

Unfortunately picked a difficult example, as the rgl one puts out two widgets (viewer and controls). Only one ought to be fluidly sized.

Also doesn't work with Quarto layout panels (we have more work t o do there generally!)


jimjam-slam:
Argh, I've based this off `main` instead of `dev`. Will clean it up next time as it's not ready to merge yet anyway! It's just meant to be https://github.com/qmd-lab/closeread/pull/181/commits/0d3042701e0b4fab620c6f8d577b2ec459292975 so far anyway.

jimjam-slam:
Close in favour of #182.

PR #182, opened 2025-07-26
WIP widget sizing work

jimjam-slam:
Still WIP! Unfortunately picked a difficult example, as the rgl one puts out two widgets (viewer and controls). Only one ought to be fluidly sized.

Also doesn't work with Quarto layout panels (we have more work to do there generally!)


jimjam-slam:
Sizing is more consistent with ce0fa54 and 7f8d0db: widgets now correctly grow to fill the space but leave room both for additional widgets (eg. the controls in the rgl example) or a figure caption (the Leaflet example).

Scenarios I haven't yet tested include:

- Multiple code cells with their own widgets in one sticky
- Quarto panel layouts (in fact, I expect that these won't work)
- Jupyter widgets (should work but I need to make the test page to check)

There's also a problem with widgets not correctly receiving mouse events. When I drag on the rgl widget, it does not respond, but the `body` gets `.leaflet-dragging`. When I drag on the Leaflet map, it does not respond.

jimjam-slam:
Have now added some tests and made some modifications to Jupyter widget treatment in 11fd774. I'm not sure why I need to force the height and width of Jupyter widgets but not htmlwidgets.

I've left out Plotly (which is in the test but not working). It doesn't seem to produce any element with a `.jupyter-widgets` class, so I'm not sure if it's actually a Jupyter widget or its own thing. That'll need more work.

To start testing Python pages reproducibly, I've added a uv project to `/docs`. Using it is:

```bash
# install the dependencies in the `pyproject.toml` and `uv.lock`
uv sync

# activate the virtual environment so quarto uses it
# (not sure if this is something quarto does automatically, but it can't hurt)
source .venv/bin/activate
```

Then render as usual. I might need to add those steps to the CI/CD actions!

jimjam-slam:
Think the `main.py` crept in from scaffolding a uv project; that can probably go!

andrewpbray:
Ah, a lock file! Whoops, I guess I should have read this before trying to render 😂 .

The sizing looks great to me. I'm getting unpredictable behavior on focus. on the R map the cursor is a hand over the whole map and strangely I cannot pan or zoom but I can (with the hand) click the + and - and they work. I'm not able to interact with the second widget at all. Similar issues with the Python one.

How is click-focus being determined in the sticky stack? I remember we bumped into this awhile back but I can't remember if/how we solved it.

jimjam-slam:
I'm getting unpredictable behavior on focus. on the R map the cursor is a hand over the whole map and strangely I cannot pan or zoom but I can (with the hand) click the + and - and they work. I'm not able to interact with the second widget at all. Similar issues with the Python one.
> 
> How is click-focus being determined in the sticky stack? I remember we bumped into this awhile back but I can't remember if/how we solved it.

Ah, yes! I think we had to fiddle with ~~`mouse-events`~~ [`pointer-events`](https://developer.mozilla.org/en-US/docs/Web/CSS/pointer-events) in the CSS a bit originally, so we might need some of that here too.

jimjam-slam:
> Ah, a lock file! Whoops, I guess I should have read this before trying to render 😂 .

Yeah, I've started using uv on some other projects and it's working well! Doesn't handle system dependencies the way conda does, but Pixi works similarly to uv and handles them via conda. (We actually had an issue with quarto-svelte recently where someone couldn't install npm on their controlled environment but was able to install both quarto and npm via pixi and render that way!) Haven't experimented with handling the R side of things with uv much, but it's enough for doing the Python stuff I think (at least for now). Might need to check whether I added it to the CI though!

PR #184, opened 2025-09-01
update banner

andrewpbray:
replaces closeread prize winners with student data scrollytelling contest


jimjam-slam (approved):
lgtm! Can't believe we left the old banner up for so long 😆

